PR #42, opened 2023-04-28
Revert on stale price borrow controller

08xmt:
Adds a check of whether a market's price feed is stale or not, and denies borrows if the price is stale.

A staleness threshold is set by the operator.


08xmt:
Consider optimizing daily borrow code to use less memory

PR #43, opened 2023-04-29
Curve helper

08xmt:


08xmt:
No need to use `msg.sender == tx.origin` as helper contract only uses `borrowOnBehalf` and contracts can't produce signatures.

PR #49, opened 2023-05-14
Improvement/add max withdraw

08xmt:
Add `withdrawMax()` and `withdrawMaxOnBehalf(...)` functions to the Market contract. These functions are necessary to allow users to deploy everything from their self-accruing personal escrows like that for xINV, and has so far been unnecessary.

The addition has the slight benefit of being slightly cheaper to execute on rollups, as the calls use less calldata than a standard withdraw.

### DIFF
--- a/src/Market.sol
+++ b/src/Market.sol
@@ -488,6 +488,15 @@ contract Market {
     }
 
     /**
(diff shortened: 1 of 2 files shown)


wavey0x (approved):

wavey0x:
lgtm

PR #57, opened 2023-08-23
Feat: wBTC market

08xmt:
* Add custom price feed using chainlink oracle for WBTC/BTC price and  chainlink oracle for BTC/USD price to compute WBTC/USD price.

The pricefeed uses the tricrypto CRV pool as a fallback oracle if the price <= max or min price of the chainlink oracle. This is to keep positions liquidateable, even in the case where WBTC may lose backing.

### DIFF
--- a/src/feeds/WbtcPriceFeed.sol
+++ b/src/feeds/WbtcPriceFeed.sol
@@ -38,7 +38,7 @@ contract WbtcPriceFeed {
         (uint80 btcUsdRoundId,int256 btcUsdPrice,uint btcUsdStartedAt,uint btcUsdUpdatedAt,uint80 btcUsdAnsweredInRound) = btcToUsd.latestRoundData();
         int wbtcUsdPrice = btcUsdPrice * 10**8 / wbtcBtcPrice;
         if(isPriceOutOfBounds(wbtcBtcPrice, wbtcToBtc)){
-            wbtcUsdPrice = wbtcToUsdFallback();
+            wbtcUsdPrice = wbtcToUsdFallbackOracle();
         }
         if(wbtcBtcUpdatedAt < btcUsdUpdatedAt){
             return (wbtcBtcRoundId, wbtcUsdPrice, wbtcBtcStartedAt, wbtcBtcUpdatedAt, wbtcBtcAnsweredInRound);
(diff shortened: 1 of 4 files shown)


src/feeds/WbtcPriceFeed.sol
```diff
@@ -0,0 +1,73 @@
...
+     * @return startedAt The timestamp when the latest round of Chainlink price feed started of the lowest last updatedAt feed
+     * @return updatedAt The lowest timestamp when either of the latest round of Chainlink price feed was updated
+     * @return answeredInRound The round ID in which the answer was computed of the lowest updatedAt feed
+     */
+    function latestRoundData() external view returns (uint80, int256, uint256, uint256, uint80){
+        (uint80 wbtcBtcRoundId,int256 wbtcBtcPrice,uint wbtcBtcStartedAt,uint wbtcBtcUpdatedAt,uint80 wbtcBtcAnsweredInRound) = wbtcToBtc.latestRoundData();
+        (uint80 btcUsdRoundId,int256 btcUsdPrice,uint btcUsdStartedAt,uint btcUsdUpdatedAt,uint80 btcUsdAnsweredInRound) = btcToUsd.latestRoundData();
+        int wbtcUsdPrice = btcUsdPrice * 10**8 / wbtcBtcPrice;
+        if(isPriceOutOfBounds(wbtcBtcPrice, wbtcToBtc)){
+            wbtcUsdPrice = wbtcToUsdFallbackOracle();
```
wavey0x, line 41: this is technically wbtcUsdtPrice, which means it could be affected by a USDT depeg.

wavey0x: Would it make sense to have some indicator in the return data to specify if the fallback oracle was used?

wavey0x (approved):

src/feeds/WbtcPriceFeed.sol
08xmt: It would make sense if our current oracle supported using this information. We're planning an update that'll introduce fallback feeds in the oracle itself, with the individual price feeds just returning whether the returned price is within safety parameters (with Chainlink being a staleness check, min and max check).

08xmt: > this is technically wbtcUsdtPrice, which means it could be affected by a USDT depeg.

It's a good point. Will incorporate a chainlink feed that reads from USDT/USD in case of using the fallback.

08xmt: > this is technically wbtcUsdtPrice, which means it could be affected by a USDT depeg.

Added a conversation from USDT to USD price in the fallback oracle in [75c5735](https://github.com/InverseFinance/FiRM/pull/57/commits/75c5735b380c9c831bdd2123afdfcf041fc1d45e). This makes the fallback reliant on Chainlink, which I'm not entirely sure is a great idea, but considering this specific fallback is to guard against a WBTC depeg event, I think it's fine for the specific feed.

I don't do any checks of the soundness of the chainlink prices returned in the fallback, as WBTC reaching max or min price is already an extreme event.

PR #62, opened 2023-10-17
Feat: return Stale price for ETH cl oracle maxAnswer

lacoop6tu:
- Return stale price instead of reverting for maxAnswer eth in fallback oracle
- Test


08xmt (commented):
lgtm

PR #63, opened 2023-10-24
Feat: ERC-4626 escrow

lacoop6tu:
- Add ERC4626 VaultHelper for:
  -  depositing underlying token into the vault and then depositing vault token into the market
  -  withdrawing vault token from market and redeeming for underlying token
- Add sFrax price feed (vault token)
- Add VaultHelper, sFrax price feed and market integration tests, 100% coverage


08xmt (commented):
lgtm

src/escrows/VaultEscrow.sol
```diff
@@ -0,0 +1,67 @@
...
+    function balance() public view returns (uint) {
+        uint vaultBalance = vault.balanceOf(address(this));
+        return vault.convertToAssets(vaultBalance);
+    }
+
+    /**
+    @notice Function called by market on deposit. Function is empty for this escrow.
+    @dev This function should remain callable by anyone to handle direct inbound transfers.
+    */
+    function onDeposit() public {
```
wavey0x, line 61: seems safe enough to have no access control here?

I'm assuming that the market will be deployed with `callOnDepositCallback = true`?

08xmt: Yep. It should be safe as there's no way to inject either the token nor the vault, as init is called on escrow creation. Will still ofc have to be set-up with the assumption of the underlying token and ERC-4626 vault being safe.

src/escrows/VaultHelper.sol
```diff
@@ -0,0 +1,79 @@
...
+    @param _vault The address of the ERC4626 vault
+    @param _market The address of the market
+    @param _token The address of the ERC20 token
+    **/
+    constructor(address _vault, address _market, address _token) {
+        vault = IERC4626(_vault);
+        market = IMarket(_market);
+        token = IERC20(_token);
+        token.approve(_vault, type(uint).max);
+        vault.approve(_market, type(uint).max);
```
08xmt, line 34: Do we need a way to refresh these approvals to prevent a flashmint griefing attack?

08xmt: Fixed in https://github.com/InverseFinance/FiRM/pull/63/commits/5ba1fef715294c0626193dccb933348884351416

PR #65, opened 2023-11-07
Feat: wstETH price feed

lacoop6tu:
Add wstETH price feed with stEth/usd chainlink oracle
Use fallback with eth/usd and stEth/eth cl oracles
Add tests with 100% coverage

### DIFF
--- a/src/feeds/WstETHPriceFeed.sol
+++ b/src/feeds/WstETHPriceFeed.sol
@@ -5,14 +5,12 @@ pragma solidity ^0.8.4;
 interface IChainlinkFeed {
     function aggregator() external view returns (address aggregator);
 
-    function decimals() external view returns (uint8);
-
     function latestRoundData()
         external
         view
         returns (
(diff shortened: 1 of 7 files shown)


src/feeds/WstETHPriceFeed.sol
```diff
@@ -0,0 +1,201 @@
...
+        ) = ethToUsd.latestRoundData();
+
+        (
+            ,
+            int256 stEthToEthPrice,
+            ,
+            uint256 updatedAtStEth,
+        ) = stEthToEth.latestRoundData();
+
+        int256 stEthToUsdPrice = ethToUsdPrice * stEthToEthPrice / 10**18;
```
wavey0x, line 160: This answer is in terms of `1e8`, while the other is `1e18`.

I think you should divide by `1e8` instead of `1e18` in order to arrive at this oracle's defined 18 decimals.

PR #67, opened 2023-12-21
Feat: DBR helper review

lacoop6tu:
Updates after review
- use DBR to check if a market is valid
- update approve function
- rm checks when claming DBR
- use ALLCAPS for constants
- update AddressZero error
- update doc


08xmt:
LGTM

PR #68, opened 2024-01-12
Feat: st-yETH market

lacoop6tu:
Add st-yETH feed
Add VaultHelper (as sFrax market)
Add tests, 100% coverage


src/escrows/VaultHelper.sol
```diff
@@ -0,0 +1,83 @@
...
+    function wrapAndDeposit(
+        address recipient,
+        uint assets
+    ) external returns (uint256 shares) {
+        if(recipient == address(0)) revert AddressZero();
+        token.safeTransferFrom(msg.sender, address(this), assets);
+        if (assets > vault.maxDeposit(address(this))) revert MaxDeposit(assets);
+        shares = vault.deposit(assets, address(this));
+        uint256 actualShares = vault.balanceOf(address(this));
+        if (shares > actualShares) revert InsufficientShares();
```
wavey0x, line 53: this is ineffective since there should never be a mismatch.
i would suggest getting rid of `actualShares` and simply ensuring `shares > 0`.
if you want a more robust defense, could consider a `minSharesOut` parameter.

src/escrows/VaultHelper.sol
```diff
@@ -0,0 +1,83 @@
...
+    @param recipient The address to receive payment from the escrow
+    @param assets The amount of ERC20 token to be transferred.
+    */
+    function wrapAndDeposit(
+        address recipient,
+        uint assets
+    ) external returns (uint256 shares) {
+        if(recipient == address(0)) revert AddressZero();
+        token.safeTransferFrom(msg.sender, address(this), assets);
+        if (assets > vault.maxDeposit(address(this))) revert MaxDeposit(assets);
```
wavey0x, line 50: while this is a good practice in general, i would note that st-yETH (specifically) has a hardcoded uncapped `maxMint` and `maxDeposit`. so this is effectively wasted gas.

wavey0x (approved):

src/escrows/VaultHelper.sol
wavey0x: for what it's worth, looking at st-yETH code, it enforces that the final 1e15 shares can never be burned. this makes the cost of an attack incredibly expensive.

PR #71, opened 2024-02-13
Add maxPrice oracle

08xmt:
Replace the dampened price logic using lowest recorded price with logic that only allows price to increase by `1000/collateralFactorBps` over a given time window.


08xmt:
Still requires a proper rewriting of the test suite

PR #74, opened 2024-05-01
feat: ERC4626 helper for ALE and market

0xtj24:
Add ERC4626 Helper
- Supports multiple vaults/markets
- Used by the ALE for leverage
- Can also be used for entering and exiting a market
- Has guardian role for pausing markets only
- Test refactor


src/util/ERC4626Helper.sol
```diff
@@ -0,0 +1,244 @@
...
+     * @notice Pause the market by setting the vault and underlying token to address(0).
+     * @dev Only callable by the governance or the guardian.
+     * @param market The address of the market to be paused.
+     */
+    function pauseMarket(address market) external onlyGuardianOrGov {
+        markets[market].vault = IERC4626(address(0));
+        markets[market].underlying = IERC20(address(0));
+        emit MarketPaused(market);
+    }
+}
```
08xmt, line 244: Rename to remove market, as that's effectively what the function does.
```suggestion
    /**
     * @notice Remove the market by setting the vault and underlying token to address(0).
     * @dev Only callable by the governance or the guardian.
     * @param market The address of the market to be removed.
     */
    function removeMarket(address market) external onlyGuardianOrGov {
        markets[market].vault = IERC4626(address(0));
        markets[market].underlying = IERC20(address(0));
        emit MarketRemoved(market);
    }
}
```

PR #76, opened 2024-06-03
Feat: Curve LP escrows

0xtj24:
WIP

### DIFF
--- a/src/feeds/DolaFraxPyUsdPriceFeed.sol
+++ b/src/feeds/DolaFraxPyUsdPriceFeed.sol
@@ -1,36 +1,7 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.4;
 
-interface IChainlinkFeed {
-    function aggregator() external view returns (address aggregator);
-
-    function decimals() external view returns (uint8);
-
-    function latestRoundData()
-        external
-        view
-        returns (
-            uint80 roundId,
-            int256 lpDollarPrice,
-            uint256 startedAt,
-            uint256 updatedAt,
-            uint80 answeredInRound
-        );
-}
-
-interface IAggregator {
-    function maxAnswer() external view returns (int192);
-
-    function minAnswer() external view returns (int192);
-}
-
-interface ICurvePool {
-    function price_oracle(uint256 k) external view returns (uint256);
-
-    function get_virtual_price() external view returns (uint256);
-
-    function ema_price() external view returns (uint256);
-}
+import "src/util/ChainlinkFeedFallbacks.sol";
 
 contract DolaFraxPyUsdPriceFeed {
     error OnlyGov();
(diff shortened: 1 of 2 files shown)


src/escrows/DolaFraxBPEscrow.sol
```diff
@@ -0,0 +1,271 @@
...
+    function moveFromYearnToConvex(
+        bool useAll
+    ) external onlyBeneficiary returns (uint256 lpAmount) {
+        lpAmount = _fullWithdrawFromYearn();
+
+        if (useAll) lpAmount = token.balanceOf(address(this));
+
+        stakedBalance += lpAmount;
+        booster.deposit(pid, lpAmount, true);
+    }
```
08xmt, line 257: Unnecessary if implementing automatic withdraws in deposit functions

src/escrows/DolaFraxBPEscrow.sol
```diff
@@ -0,0 +1,271 @@
...
+            yearn.balanceOf(address(this)),
+            address(this)
+        );
+    }
+
+    function _fullWithdrawFromConvex() internal returns (uint256 lpAmount) {
+        lpAmount = stakedBalance;
+        stakedBalance = 0;
+        rewardPool.withdrawAndUnwrap(lpAmount, false);
+    }
```
08xmt, line 270: Move logic to withdraw functions as withdraw functions are currently just empty wrappers for these.

08xmt (changes requested):

src/escrows/DolaFraxBPEscrow.sol
08xmt: Imo having the user deposit all into the strategy they decide on is the way to go, but worth bringing up with others if they think it's better to let the user choose.

src/feeds/DolaFraxPyUsdPriceFeed.sol
08xmt, line 1: I think we can generalize and unbundle these very long price feeds into multiple smaller ones, and then combine them into what we need the end result to be.

Maybe there's some hidden complexity I don't see, but it seems like we should be able to do the following:

1. Create a CRV LP price feed which takes two price feeds in it's constructor.
2. Create a ChainlinkFallback pricefeed, which has the normal price feed functions along with the following functions:
- `function isPriceStale() public returns (bool)`
- `function hasFallback() public returns (bool)`
- `function getFallback() public returns (ChainlinkFallbackPriceFeed)`
And have the fallback logic built into it. In turn the Fallback can then have its own fallbacks if desired.
3. Create a CurveEMAPriceFeed, that takes a curve pool and chainlink feed in its constructor, serving as a generalized EMA price feed.

08xmt: See https://github.com/InverseFinance/FiRM/pull/77

PR #78, opened 2024-06-17
Refactor: use generalized feeds for DolaFraxBP and DolaFraxPyUSD Feeds

0xtj24:
WIP
- Use generalization for DolaFraxBP, DolaFraxPyUsd and crvUSD/Dola
- Add Base Feeds
- Use base test


src/feeds/ChainlinkEmaPriceFeed.sol
```diff
@@ -0,0 +1,149 @@
+pragma solidity ^0.8.20;
+
+import "src/util/FeedLib.sol";
+
+// Combined Chainlink EMA Price Feed, allows for additional fallback to be set
+
+contract ChainlinkEmaPriceFeed {
+    IChainlinkFeed public immutable assetToUsd;
+    IChainlinkFeed public assetToUsdFallback;
```
08xmt, line 9: Can use `ChainlinkBasePriceFeed` for `assetToUsd` and get rid of the `assetToUsdFallback`

src/feeds/ChainlinkEmaPriceFeed.sol
```diff
@@ -0,0 +1,149 @@
...
+    ICurvePool public immutable curvePool;
+    address public owner;
+    address public pendingOwner;
+    uint256 public assetToUsdHeartbeat;
```
08xmt, line 13: Unnecessary if using `ChainlinkBasePriceFeed` as `assetToUsd` feed

src/feeds/ChainlinkEmaPriceFeed.sol
```diff
@@ -0,0 +1,149 @@
...
+        returns (
+            uint80 roundId,
+            int256 usdPrice,
+            uint256 startedAt,
+            uint256 updatedAt,
+            uint80 answeredInRound
+        )
+    {
+        if (isPriceStale()) {
+            if (hasFallback()) {
```
08xmt, line 63: No need for staleness checks if using `ChainlinkBasePriceFeed`

src/feeds/ChainlinkEmaPriceFeed.sol
08xmt, line 1: Can be simplified greatly by forcing it to use the `ChainlinkBasePriceFeed`

src/feeds/ChainlinkPriceOracleStablePriceFeed.sol
08xmt, line 1: Can be simplified greatly by forcing it to use the `ChainlinkBasePriceFeed`

08xmt (changes requested):

src/feeds/ChainlinkEmaPriceFeed.sol
08xmt, line 1: Should always use `PriceOracle` implementation over this one. So file is unnecessary

08xmt: Shouldn't we still avoid any use of `EMA_PRICE()`? Or is it fine in this case?

src/feeds/ChainlinkCurveAssetFeed.sol
```diff
@@ -4,35 +4,25 @@ import {ICurvePool} from "src/interfaces/ICurvePool.sol";
 import {IChainlinkBasePriceFeed} from "src/interfaces/IChainlinkFeed.sol";
 
 // Combined Chainlink and Curve price_oracle, allows for additional fallbacks to be set via ChainlinkBasePriceFeed
-
-contract ChainlinkPriceOracleStablePriceFeed {
+/// @dev This implementation is for Curve pools when Target is the coin at index 0 in Curve pool `coins` array
```
08xmt, line 7: Can we always assume that the target will be index 0 or should we generalize it to take coin index parameters?

src/feeds/ChainlinkCurveTargetFeed.sol
```diff
@@ -4,36 +4,25 @@ import {ICurvePool} from "src/interfaces/ICurvePool.sol";
 import {IChainlinkBasePriceFeed} from "src/interfaces/IChainlinkFeed.sol";
 
 // Combined Chainlink and Curve price_oracle, allows for additional fallback to be set via ChainlinkBasePriceFeed
-
-contract ChainlinkPriceOracleFeed {
+/// @dev This implementation is for Curve pools when Asset is the coin at index 0 in Curve pool `coins` array
```
08xmt, line 7: Can we always assume that the target will be index 0 or should we generalize it to take coin index parameters?

src/feeds/ChainlinkPriceOracleFeed.sol
```diff
@@ -0,0 +1,98 @@
...
+import {ICurvePool} from "src/interfaces/ICurvePool.sol";
+import {IChainlinkBasePriceFeed} from "src/interfaces/IChainlinkFeed.sol";
+
+// Combined Chainlink and Curve price_oracle, allows for additional fallback to be set via ChainlinkBasePriceFeed
+
+contract ChainlinkPriceOracleFeed {
+    int256 public constant SCALE = 1e18;
+    IChainlinkBasePriceFeed public immutable assetToUsd;
+    ICurvePool public immutable curvePool;
+    uint256 public immutable targetK;
```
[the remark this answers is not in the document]
08xmt: Suggest we give contract a `targetIndex` param in constructor, so contract can be used even if the target coin isn't index 0

src/feeds/CurveLPPessimisticFeed.sol
```diff
@@ -0,0 +1,87 @@
...
+            uint updatedAtCoin2,
+            uint80 answeredInRoundCoin2
+        ) = coin2Feed.latestRoundData();
+
+        int256 minLpUsdPrice;
+
+        // If coin1 price is lower than coin2 price, use coin1 price
+        if (
+            (usdPriceCoin1 < usdPriceCoin2 && updatedAt > 0) ||
+            usdPriceCoin2 == 0
```
08xmt, line 53: Why is this check necessary?

08xmt (approved):
Good work

src/escrows/DolaFraxPyUSDEscrow.sol
08xmt, line 1: A bit awkward that there's no way to generalize this escrow. Would it be possible to feed it the contract addresses in another way than hard coding them?

08xmt: Could make an abstract version that doesn't implement the addresses, but implements all the logic.

src/escrows/DolaFraxPyUSDEscrow.sol
```diff
@@ -0,0 +1,239 @@
...
+            // If there are enough tokens in Yearn, withdraw the amount from Yearn
+            uint256 withdrawAmount = YearnVaultV2Helper.assetToCollateral(
+                yearn,
+                missingAmount
+            );
+            // Withdraw from Yearn
+            uint256 assetAmount = yearn.withdraw(withdrawAmount, address(this));
+            // Transfer the amount to the recipient
+            token.safeTransfer(recipient, assetAmount + tokenBal);
+        }
```
08xmt, line 105: This should have terminating logic that reverts on no transfer. It's very dangerous for the pay function to be able to execute without transferring anything and without reverting.

src/escrows/DolaFraxPyUSDEscrow.sol
```diff
@@ -0,0 +1,239 @@
...
+    */
+    function balance() public view returns (uint) {
+        return
+            stakedBalance +
+            YearnVaultV2Helper.collateralToAsset(
+                yearn,
+                yearn.balanceOf(address(this))
+            ) +
+            token.balanceOf(address(this));
+    }
```
08xmt, line 120: This balance function can cause problems, as the pay function currently can only transfer `max(stakedBalance + tokenBal, yearnBalance + tokenBal)`

08xmt: In the worst case scenario it could be used to trigger a `pay(...)` call that results in no tokens being transferred, which in turn could potentially be used to erase your own debt through self-liquidations without losing assets.

src/escrows/DolaFraxPyUSDEscrow.sol
```diff
@@ -0,0 +1,239 @@
...
+    IYearnVaultV2 public constant yearn =
+        IYearnVaultV2(0x5b737CC835c29c493845353e0399B408Fbd7105D);
+    IERC20 public constant cvx =
+        IERC20(0x4e3FBD56CD56c3e72c1403e103b45Db9da5B9D2B);
+    IERC20 public constant crv =
+        IERC20(0xD533a949740bb3306d119CC777fa900bA034cd52);
+
+    address public market;
+    IERC20 public token;
+    uint256 public stakedBalance;
```
08xmt, line 39: It may be worth dropping the `stakedBalance` counter if it's possible to deposit to convex on behalf of the contract. Otherwise adding a function call that allows reading the actual balance of the staking contract and updating `stakedBalance` accordingly may be a good addition.

08xmt (changes requested):

src/escrows/LPCurveYearnV2Escrow.sol
```diff
@@ -73,31 +91,51 @@ contract DolaFraxBPEscrow {
...
-            uint256 assetAmount = yearn.withdraw(withdrawAmount, address(this));
-            // Transfer the amount to the recipient
-            token.safeTransfer(recipient, assetAmount + tokenBal);
+            yearn.withdraw(collateralAmount, address(this));
+
+            uint256 lpToPay = token.balanceOf(address(this));
+            if (lpToPay != amount) {
+                _ensureLimitsOrRevert(lpToPay, amount);
+                amount = lpToPay;
+            }
```
08xmt, line 136: 1. Is this check necessary? Looks to me as it'll revert on transfer if `lpToPay < amount`, and that there's no real issue if `lpToPay > weiDelta`.
2. There's no reason to pay the `weiDelta` as long as we have enough lp tokens to transfer `amount`.

08xmt (commented):
Good work

08xmt (approved):

src/feeds/CurveLPPessimisticFeed.sol
```diff
@@ -0,0 +1,84 @@
...
+            int256 usdPriceCoin2,
+            uint startedAtCoin2,
+            uint updatedAtCoin2,
+            uint80 answeredInRoundCoin2
+        ) = coin2Feed.latestRoundData();
+
+        int256 minLpUsdPrice;
+
+        // If coin1 price is lower than coin2 price, use coin1 price
+        if ((usdPriceCoin1 < usdPriceCoin2 && updatedAt > 0)) {
```
wavey0x, line 51: Are you sure that, in theory, pool shares (LPs) always get minted at a rate that is <= 1:1 ?

i.e. imagine if some pool type mints 100 shares to someone who deposits 1USDC:1DAI ($2 value). the feed would show ~$100 value.

i'm not aware of Curve pools where this is the case, but worth checking since there's so many types now. Tricrypto, V2, volatile pairs, etc.

src/escrows/LPCurveYearnV2Escrow.sol
```diff
@@ -73,31 +90,43 @@ contract DolaFraxBPEscrow {
...
-            ) >= missingAmount
-        ) {
-            // If there are enough tokens in Yearn, withdraw the amount from Yearn
-            uint256 withdrawAmount = YearnVaultV2Helper.assetToCollateral(
+            return;
+        }
+
+        uint256 missingAmount = amount - tokenBal;
+        uint256 convexBalance = rewardPool.balanceOf(address(this));
+        if (convexBalance > 0 && missingAmount > 0) {
```
wavey0x, line 101: the check on `missingAmount > 0` is redundant. if it were 0 , line would be unreachable due to early return above.

src/escrows/LPCurveYearnV2Escrow.sol
```diff
@@ -73,31 +90,43 @@ contract DolaFraxBPEscrow {
...
+            else
+                collateralAmount = YearnVaultV2Helper.assetToCollateral(
+                    yearn,
+                    withdrawAmount + weiDelta
+                );
             // Withdraw from Yearn
-            uint256 assetAmount = yearn.withdraw(withdrawAmount, address(this));
-            // Transfer the amount to the recipient
-            token.safeTransfer(recipient, assetAmount + tokenBal);
+            yearn.withdraw(collateralAmount, address(this));
```
wavey0x, line 127: Suggestion: in an extremely unlikely scenario, there can be a loss on a yearn vault. If a loss exceeds 0.01% (default in the vault), the escrow's position will be stuck.

i would suggest replacing your current interface with one that includes the optional `maxLoss` parameter. and allow users to adjust this value. 
Reference on vault interface is here...

https://github.com/yearn/yearn-vaults/blob/97ca1b2e4fcf20f4be0ff456dabd020bfeb6697b/contracts/Vault.vy#L1033

wavey0x: Same for line 247.

src/escrows/LPCurveYearnV2Escrow.sol
```diff
@@ -193,7 +230,8 @@ contract DolaFraxBPEscrow {
      * @dev Cannot deposit if there are Convex tokens in the escrow (only 1 strategy at a time)
      */
     function depositToYearn() external onlyBeneficiary {
-        if (stakedBalance > 0) withdrawFromConvex();
+        uint256 convexBalance = rewardPool.balanceOf(address(this));
+        if (convexBalance > 0) withdrawFromConvex();
         yearn.deposit(token.balanceOf(address(this)), address(this));
```
wavey0x, line 235: Optional:
best practice for vault integrations is to check the `vault.availableDepositLimit()` such that you only deposit what the vault can allow, and not run into a revert.

In your case, nothing bad can happen. Worst case is a revert, without blocking critical code paths. Yearn's Curve LP vaults typically all have extremely high deposit limits that never get reached.

reference:
https://github.com/yearn/yearn-vaults/blob/97ca1b2e4fcf20f4be0ff456dabd020bfeb6697b/contracts/Vault.vy#L1605

wavey0x (approved):

wavey0x:
LGTM

src/feeds/CurveLPPessimisticFeed.sol
wavey0x: sounds good. as far as i know, you are correct with curve mechanics. what troubles me about the assumption is seeing velodrome/aerodrome pools which issue LPs at distinctly different prices than 1:1. but should be good as long as things are checked case by case.

refer to this pool which has $800k TVL, and each 1 LP is priced at $2,000,156,024,660.25
https://basescan.org/token/0x27a8Afa3Bd49406e48a074350fB7b2020c43B2bD

PR #82, opened 2024-08-04
Feat/ALE update and CurveDolaLPHelper

0xtj24:
- Integration with FlashMinter so ALE doesn’t require a minter role.
- Optionality to use or not the exchangeProxy (required for CurveLP ALE actions)
- Added support to `depositAndLeverage` with both collateral and underlying (DOLA in case of plain CurveLP or YearnV2 Dola LP Vaults)
- Add CurveDolaLPHelper for ALE and markets
- Add tests


src/util/ALE.sol
```diff
@@ -0,0 +1,770 @@
...
+            // Approve sellToken for spender
+            sellToken.approve(_spender, 0);
+            sellToken.approve(_spender, _collateralAmount);
+            (bool success, ) = exchangeProxy.call{value: msg.value}(
+                _swapCallData
+            );
+            if (!success) revert SwapFailed();
+        }
+
+        if (address(markets[_market].helper) == address(0)) {
```
08xmt, line 588: What should happen if there is a helper contract? Helper contract pays back excess?

wavey0x (approved):

wavey0x (approved):

PR #88, opened 2024-09-25
Feat: update yvyCRV helper contract

0xtj24:
- Use YearnVaultV2Helper.sol for collateral conversion estimation
- Update naming to yvyCRV
- Use Sweepable.sol instead of Ownable.sol
- Update test


08xmt (approved):
lgtm, only thing is whether it'd be worth it to generalize the contract.

test/util/aleTests/ALEyvyCRVHelperForkTest.t.sol
```diff
@@ -168,40 +115,43 @@ contract ALEHelperForkTest is Test {
...
             stYCRVDeposit + collateralToSwap,
             1
         );
     }
 
-    function test_leveragePosition() public {
-        // vm.assume(styCRVAmount < 7900 ether);
-        // vm.assume(styCRVAmount > 0.00000001 ether);
+    function test_leveragePosition(uint256 styCRVAmount) public {
+        vm.assume(styCRVAmount < 7900 ether);
```
08xmt, line 142: Why is this fuzz test limited to such a relatively small amount?

test/util/aleTests/ALEyvyCRVHelperForkTest.t.sol
```diff
@@ -267,20 +218,20 @@ contract ALEHelperForkTest is Test {
         assertEq(DOLA.balanceOf(userPk), 0);
```
08xmt, line 218: Consider adding assert that checks `.balance()` function of escrow, as that's arguably more important than the `yvyCRV.balanceOf(escrowAddr)` number, even though the two *should* be equivalent.

test/util/aleTests/ALEyvyCRVHelperForkTest.t.sol
```diff
@@ -267,20 +218,20 @@ contract ALEHelperForkTest is Test {
...
     }
 
-    function test_depositAndLeveragePosition() public {
-        // vm.assume(styCRVAmount < 7900 ether);
-        // vm.assume(styCRVAmount > 0.00000001 ether);
+    function test_depositAndLeveragePosition(uint256 styCRVAmount) public {
+        vm.assume(styCRVAmount < 7900 ether);
```
08xmt, line 222: Why is styCRVAmount so low in this fuzz test?

test/util/aleTests/ALEyvyCRVHelperForkTest.t.sol
```diff
@@ -337,9 +288,10 @@ contract ALEHelperForkTest is Test {
 
         // Balance in escrow is equal to the collateral deposited + the extra collateral swapped from the leverage
         assertApproxEqAbs(
```
08xmt, line 290: Consider adding asserts that check balance of escrow and that `maxBorrowAmount` hasn't been exceeded.

test/util/aleTests/ALEyvyCRVHelperForkTest.t.sol
08xmt: Is the best approach to use the balance of an EOA as the fuzz limit? The fuzz test should not be limited by what some arbitrary holder has, but rather by constraints that make sense in terms of the contract code. Is there a reason for not using something like `vm.deal(styCrv, testUser, fuzz_amount )`?

PR #90, opened 2024-10-03
Add marketv2

08xmt:


src/MarketV2.sol
```diff
@@ -0,0 +1,679 @@
...
+// Caution. We assume all failed transfers cause reverts and ignore the returned bool.
+interface IOracle {
+    function getPrice(address,uint) external returns (uint);
+    function viewPrice(address,uint) external view returns (uint);
+}
+
+interface IEscrow {
+    function initialize(IERC20 _token, address beneficiary) external;
+    function onDeposit() external;
+    function onDepositCallBack() external returns(uint);
```
nourharidy, line 14: What's this function?

src/MarketV2.sol
```diff
@@ -0,0 +1,679 @@
...
+
+interface IBorrowController {
+    function borrowAllowed(address msgSender, address borrower, uint amount) external returns (bool);
+    function onRepay(uint amount) external;
+}
+
+interface IDebtManager {
+	function debt(address borrower) external view returns(uint);
+	function totalDebt() external view returns(uint);
+	function dbrDeficit(address borrower) external view returns(uint);
```
nourharidy, line 36: Debt managers shouldn't have to deal with DBR and therefore this function shouldn't be on the interface. Maybe it can be replaced with a boolean `inDeficit(address) returns (bool)` function which would be universal whether using DBR or not.

src/MarketV2.sol
```diff
@@ -0,0 +1,679 @@
...
+	function increaseDebt(address borrower, uint amount) external;
+	function decreaseDebt(address borrower, uint amount) external returns(uint);
+	function replenish(address borrower, uint amount) external;
```
nourharidy, line 39: Seems unused? imo replenishments can happen through the debt manager contract directly. Doesn't have to be done through the market anyway.

src/MarketV2.sol
```diff
@@ -0,0 +1,679 @@
...
+    function DOMAIN_SEPARATOR() public view virtual returns (bytes32) {
+        return block.chainid == INITIAL_CHAIN_ID ? INITIAL_DOMAIN_SEPARATOR : computeDomainSeparator();
+    }
+
+    function computeDomainSeparator() internal view virtual returns (bytes32) {
+        return
+            keccak256(
+                abi.encode(
+                    keccak256("EIP712Domain(string name,string version,uint256 chainId,address verifyingContract)"),
+                    keccak256(bytes("DBR MARKET")),
```
nourharidy, line 121:
```suggestion
                    keccak256(bytes("FiRM MARKET")),
```

src/MarketV2.sol
```diff
@@ -0,0 +1,679 @@
...
+    @notice Internal function for withdrawing from the escrow
+    @dev The internal function is shared by the withdraw function and withdrawOnBehalf function
+    @param from The address owning the escrow to withdraw from.
+    @param to The address receiving the tokens
+    @param amount The amount being withdrawn.
+    */
+    function withdrawInternal(address from, address to, uint amount) internal {
+        uint limit = getWithdrawalLimitInternal(from);
+        require(limit >= amount, "Insufficient withdrawal limit");
+        require(dbr.deficitOf(from) == 0, "Can't withdraw with DBR deficit");
```
nourharidy, line 500: Does this rule still apply? I think it should be replaced with the user's debtManager's deficit instead, no?

PR #91, opened 2024-10-03
Add initial variableDebtManager and HookAMM

08xmt:
Variable rate market

### DIFF
--- a/src/debtManagers/HookAMM.sol
+++ b/src/debtManagers/HookAMM.sol
@@ -1,0 +1,201 @@
+// SPDX-License-Identifier: MIT License
+pragma solidity 0.8.20;
+
+interface IERC20 {
+    function transfer(address, uint) external returns (bool);
+    function transferFrom(address, address, uint) external returns (bool);
+    function balanceOf(address) external view returns (uint);
+}
+
+interface Mintable is IERC20 {
+    function burn(uint) external;
+    function mint(address, uint) external;
+}
+
+interface IVariableDebtManager {
+    function buyHook() external returns (bool);
+}
+
+/**
+ * @title HookAMM
+ * @dev Limited access xy=k AMM designed to be maximally manipulation resistant on behalf of the DebtManager and FiRMv2 markets
+*/
+contract HookAMM {
+    
+    Mintable public immutable dbr;
+    IERC20 public immutable dola;
+    IVariableDebtManager public variableDebtManager;
+    address public gov;
+    address public pendingGov;
+    address public feeRecipient;
+    uint public prevK;
+    uint public targetK;
+    uint public lastKUpdate;
+    uint public maxDbrPrice;
+    uint public dbrBuyFee;
+    uint public feesAccrued;
+
+    error Invariant();
+
+    /**
+     * @dev Constructor for sDola contract.
+     * WARNING: MIN_SHARES will always be unwithdrawable from the vault. Deployer should deposit enough to mint MIN_SHARES to avoid causing user grief.
+     * @param _dola Address of the DOLA token.
+     * @param _gov Address of the governance.
+     * @param _K Initial value for the K variable used in calculations.
+     */
+    constructor(
+        address _dola,
+        address _dbr,
+        address _gov,
+        address _feeRecipient,
+        uint _K
+    ) {
+        require(_K > 0, "_K must be positive");
+        dbr = Mintable(_dbr);
+        dola = IERC20(_dola);
+        gov = _gov;
+        feeRecipient = _feeRecipient;
+        targetK = _K;
+    }
+
+    modifier onlyGov() {
+        require(msg.sender == gov, "ONLY GOV");
+        _;
+    }
+
+    modifier buyHook() {
+        variableDebtManager.buyHook();
+        _;
+    }
+
+    /**
+     * @dev Returns the current value of K, which is a weighted average between prevK and targetK.
+     * @return The current value of K.
+     */
+    function getK() public view returns (uint) {
+        uint duration = 7 days;
+        uint timeElapsed = block.timestamp - lastKUpdate;
+        if(timeElapsed > duration) {
+            return targetK;
+        }
+        uint targetWeight = timeElapsed;
+        uint prevWeight = duration - timeElapsed;
+        return (prevK * prevWeight + targetK * targetWeight) / duration;
+    }
+
+    /**
+     * @dev Calculates the DOLA reserve based on the current DBR reserve.
+     * @return The calculated DOLA reserve.
+     */
+    function getDolaReserve() public view returns (uint) {
+        return dola.balanceOf(address(this));
+    }
+
+    /**
+     * @dev Returns the current DBR reserve as the sum of dbr balance and claimable dbr
+     * @return The current DBR reserve.
+     */
+    function getDbrReserve() public view returns (uint) {
+        return getK() / getDolaReserve();
+    }
+
+    /**
+     * @dev Sets a new target K value.
+     * @param _K The new target K value.
+     */
+    function setTargetK(uint _K) external onlyGov {
+        require(_K > getDbrReserve(), "K must be larger than dbr reserve");
+        prevK = getK();
+        targetK = _K;
+        lastKUpdate = block.timestamp;
+        emit SetTargetK(_K);
+    }
+
+    /**
+     * @dev Allows users to buy DBR with DOLA.
+     * WARNING: Never expose this directly to a UI as it's likely to cause a loss unless a transaction is executed immediately.
+     * Instead use the sDolaHelper function or custom smart contract code.
+     * @param exactDolaIn The exact amount of DOLA to spend.
+     * @param exactDbrOut The exact amount of DBR to receive.
+     * @param to The address that will receive the DBR.
+     */
+    function buyDBR(uint exactDolaIn, uint exactDbrOut, address to) external buyHook {
+        require(to != address(0), "Zero address");
+        _invariantCheck(exactDolaIn, exactDbrOut, getDbrReserve());
+        dola.transferFrom(msg.sender, address(this), exactDolaIn);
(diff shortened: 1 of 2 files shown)


src/debtManagers/HookAMM.sol
```diff
@@ -0,0 +1,238 @@
...
+        maxDolaPrice = _maxDolaPrice;
+    }
+
+    function setMinDolaPrice(uint _minDolaPrice) external onlyGov {
+        require(maxDolaPrice >= _minDolaPrice, "Max DOLA price must be higher or equal to min DOLA price");
+        minDolaPrice = _minDolaPrice;
+    }
+
+    function getVirtualDolaReserves() internal view returns(uint) {
+        return sqrt(1e18 * getK() / minDolaPrice);
```
nourharidy, line 127: Can you explain what's going on here? Why is minDolaPrice used here? and how does this square root result in dola reserves? Shouldn't we be dividing K by DBR reserve to get DOLA reserves?

src/debtManagers/HookAMM.sol
```diff
@@ -0,0 +1,238 @@
...
+        uint targetWeight = timeElapsed;
+        uint prevWeight = duration - timeElapsed;
+        return (prevK * prevWeight + targetK * targetWeight) / duration;
+    }
+
+    /**
+     * @dev Calculates the DOLA reserve based on the current DBR reserve.
+     * @return The calculated DOLA reserve.
+     */
+    function getDolaReserve() public view returns (uint) {
```
nourharidy, line 92: DBR reserve is not used here as claimed in the natspec comment. Can you explain how dola reserve should be calculated in the context of this AMM?

src/debtManagers/HookAMM.sol
```diff
@@ -0,0 +1,238 @@
...
+        uint dolaReserve = getDolaReserve();
+        uint dbrReserve = k / dolaReserve;
+        _invariantCheck(dbrReserve - exactDbrIn, dolaReserve - exactDolaOut, getDolaReserve());
+        dbr.transferFrom(msg.sender, address(this), exactDolaOut);
+        dola.transfer(to, exactDbrIn);
+        emit BuyDOLA(msg.sender, to, exactDbrIn, exactDolaOut);
+    }
+
+    function _invariantCheck(uint newDolaReserve, uint newDbrReserve, uint k) internal view {
+        if(newDolaReserve * newDbrReserve < k){
```
nourharidy, line 170: Shouldn't it be `>` rather than `<`?

src/debtManagers/HookAMM.sol
```diff
@@ -0,0 +1,238 @@
...
+            if(1e18 * newDbrReserve / newDolaReserve > maxDolaPrice){
+                uint maxDolaReserve = sqrt(1e18 * k / maxDolaPrice);
+                uint maxDbr = k / maxDolaReserve;
+                uint excessDola = newDolaReserve - maxDolaReserve;
+                uint excessDbr = newDbrReserve - maxDbr;
+                if(1e18 * excessDbr / excessDola >= maxDolaPrice) revert Invariant();
```
nourharidy, line 176: Why are we enforcing price invariants when the price is determined by our own AMM? This will essentially halt trading beyond price constraints while the intended behavior would be to continue trading at the price constraints. I think maybe the invariant design might be inappropriate here and instead we could opt for a simple price function that derives price from xy=k up until the constraints then returns fixed prices beyond them.

src/debtManagers/HookAMM.sol
```diff
@@ -0,0 +1,238 @@
...
+     */
+    function buyDBR(uint exactDolaIn, uint exactDbrOut, address to) external buyHook {
+        require(to != address(0), "Zero address");
+        _invariantCheck(exactDolaIn, exactDbrOut, getDbrReserve());
+        dola.transferFrom(msg.sender, address(this), exactDolaIn);
+        uint dbrBal = dbr.balanceOf(address(this));
+        if(exactDbrOut > dbrBal)
+            dbr.mint(address(this), exactDbrOut - dbrBal);
+        uint exactDbrOutAfterFee = exactDbrOut * (1e18 - dbrBuyFee) / 1e18;
+        feesAccrued += exactDbrOut - exactDbrOutAfterFee;
```
nourharidy, line 146: Why mint the fee at all? Instead we should deduct it from the minted amount (or burn it from balance) and remove the feeRecipient role entirely.

src/debtManagers/HookAMM.sol
```diff
@@ -0,0 +1,238 @@
...
+     * @param token The address of the ERC20 token to sweep.
+     * @param amount The amount of tokens to sweep.
+     * @param to The recipient address of the swept tokens.
+     */
+    function sweep(address token, uint amount, address to) public onlyGov {
+        require(address(dbr) != token, "Not authorized");
+        IERC20(token).transfer(to, amount);
+    }
+
+    function harvest() public {
```
nourharidy, line 212: Consider removing the feeRecipient role and removing this function. We can simply burn the excess DBR fees

src/debtManagers/VariableDebtManager.sol
```diff
@@ -0,0 +1,132 @@
...
+
+interface IDebtManager {
+    function increaseDebt(address user, uint amount) external;
+    function decreaseDebt(address user, uint amount) external returns(uint);
+    function debt(address market, address user) external view returns (uint);
+    function marketDebt(address market) external view returns(uint);
+}
+
+interface IDbrAMM {
+    function burnDbr(uint exactDolaIn, uint exactDbrBurn) external returns (uint dolaIn);
```
nourharidy, line 12: Potential case mismatch. It's called `burnDBR` on the HookAMM. Also it doesn't have a return value. Please double check interfaces.

src/debtManagers/VariableDebtManager.sol
```diff
@@ -0,0 +1,132 @@
...
+    function mint(address, uint) external;
+}
+
+
+contract VariableDebtManager is IDebtManager {
+
+    IDbrAMM public immutable amm; //TODO: Doesn't necessarily have to be immutable. Can consider making mutable.
+    IDBR public immutable dbr;
+    IDOLA public immutable dola;
+    IHelper public helper; //Should this functionality be a part of the AMM?
```
nourharidy, line 34: If we dump the invariant model in favor of a simple price function then yes a helper wouldn't be needed

src/debtManagers/VariableDebtManager.sol
```diff
@@ -0,0 +1,132 @@
...
+    
+    //Should be called when switching fixed rate debt to variable debt and when borrowing variable debt
+    function increaseDebt(address user, uint additionalDebt) external onlyMarket updateDebt {
+        address market = msg.sender;
+        totalDebt += additionalDebt;
+        uint additionalDebtShares;
+        if(totalDebt == 0){
+            additionalDebtShares = additionalDebt * MANTISSA; //Minting a high amount of initial debt shares, as debt per share will increase exponentially over the lifetime of the contract
+        } else {
+            additionalDebtShares = additionalDebt * totalDebtShares / totalDebt; //TODO: Consider rounding up in favour of other users
```
nourharidy, line 72: We can simply add a `+ 1` instead of more complex rounding up, wdyt?

src/debtManagers/VariableDebtManager.sol
```diff
@@ -0,0 +1,132 @@
...
+        if(userDebt <= amount){
+            totalDebtShares -= debtShares[market][user];
+            marketDebtShares[market] -= debtShares[market][user];
+            debtShares[market][user] = 0;
+            totalDebt -= userDebt;
+            return userDebt;
+        } else {
+            uint removedDebtShares = totalDebtShares * amount / totalDebt;
+            totalDebt -= amount;
+            totalDebtShares -= removedDebtShares; //TODO: Make sure this doesn't underflow
```
nourharidy, line 92: It's impossible to underflow given the if condition above, no? It'll always be less than totalDebtShares

src/debtManagers/VariableDebtManager.sol
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    function _burnDbrDeficit() internal {
+        if(lastUpdate < block.timestamp){
+            uint _dbrDeficit = dbrDeficit();
+            uint dolaNeeded = helper.dolaNeededForDbr(_dbrDeficit);
+            totalDebt += dolaNeeded;
+            lastUpdate = block.timestamp;
+            dola.mint(address(this), dolaNeeded);
+            amm.burnDbr(dolaNeeded, _dbrDeficit);
```
nourharidy, line 110: Careful of the interface mismatch. Should be `burnDBR`

src/debtManagers/VariableDebtManager.sol
```diff
@@ -0,0 +1,132 @@
...
+        _burnDbrDeficit();
+    }
+
+    function _burnDbrDeficit() internal {
+        if(lastUpdate < block.timestamp){
+            uint _dbrDeficit = dbrDeficit();
+            uint dolaNeeded = helper.dolaNeededForDbr(_dbrDeficit);
+            totalDebt += dolaNeeded;
+            lastUpdate = block.timestamp;
+            dola.mint(address(this), dolaNeeded);
```
nourharidy, line 109: There's no approval to the AMM, how is it expected to pull the DOLAs from this contract?

Also I'm not sure debt manager's should each have DOLA minting rights. The risk seems disproprotionately high. Either we should pre-deposit DOLAs in it or there should be an intermediate middleware than restricts infinite minting. I prefer the first option imo

src/debtManagers/VariableDebtManager.sol
```diff
@@ -0,0 +1,132 @@
...
+            amm.burnDbr(dolaNeeded, _dbrDeficit);
+        }
+    }
+
+    function dbrDeficit() public view returns (uint){
```
nourharidy, line 114: In the market contract on the #56 branch, the interface of this function has an address borrower param. Seems to mismatch here.

PR #92, opened 2024-10-04
Add initial liquidation update

08xmt:


webmass:
Suggestion: make the V2 contract more backward-compatible regarding interface with V1 to would avoid some third-party integrations to break and keep consistency when possible, I'm thinking of `collateralFactorBps` and `borrowPaused`.

08xmt:
> Suggestion: make the V2 contract more backward-compatible regarding interface with V1 to would avoid some third-party integrations to break and keep consistency when possible, I'm thinking of `collateralFactorBps` and `borrowPaused`.

Will add getter functions to preserve that backwards compatibility where possible.

PR #96, opened 2024-11-09
Feat: Linear reset borrow controller

08xmt:
Add BorrowControllerV2 contract that updates the borrowlimit on an ongoing basis rather than at midnight every day.


src/BorrowController.sol
```diff
@@ -125,22 +120,46 @@ contract BorrowController {
...
+        // cap the limit to the maximum daily limit
+        remainingDailyBorrowLimit[market]= newLimit < dailyLimit ? newLimit : dailyLimit;
+        lastDailyBorrowLimitUpdate[msg.sender] = block.timestamp;
+    }
+
+    /**
+     * @notice Returns the available borrow limit of a market
+     * @param market The address of the market to return the available borrow limt
+     */
+    function availableBorrowLimit(address market) public view returns(uint) {
```
nourharidy, line 156: Instead of duplicating code and risking inconsistency, I would suggest calling this function in `updateDailyBorrowLimit`.

src/BorrowController.sol
```diff
@@ -125,22 +120,46 @@ contract BorrowController {
...
+            uint dailyLimit = dailyBorrowLimit[msg.sender];
+            remainingDailyBorrowLimit[msg.sender] = newLimit < dailyLimit ? newLimit : dailyLimit;
         }
     }
 
+    /**
+     * @notice Updates the daily borrow limit for the market
+     * @param market The address of the market which limit will be updated
+     */
+    function updateDailyBorrowLimit(address market) internal {
```
nourharidy, line 140: Maybe if reduce most of this function's code by reusing `availableBorrowLimit()`, we can get rid of this function entirely and put its code inline within `onRepay` and `onBorrow`. This way we can avoid double writing to `remainingDailyBorrowLimit` as you wanted.

08xmt: Yeah, definitely the way to go. Added in https://github.com/InverseFinance/FiRM/pull/96/commits/a6816168f19cdad2a5747f14745f8c5be3116eb5

src/BorrowController.sol
08xmt: Removed `updateDailyBorrowLimit` in https://github.com/InverseFinance/FiRM/pull/96/commits/a6816168f19cdad2a5747f14745f8c5be3116eb5 and replaced with in-line code in `onRepay` and `onBorrow`

src/BorrowController.sol
```diff
@@ -29,7 +29,8 @@ contract BorrowController {
     mapping(address => bool) public contractAllowlist;
     mapping(address => uint) public dailyLimits;
     mapping(address => uint) public stalenessThreshold;
-    mapping(address => mapping(uint => uint)) public dailyBorrows;
+    mapping(address => uint) public remainingDailyBorrowLimit;
+    mapping(address => uint) public lastDailyBorrowLimitUpdate;
```
wavey0x, line 33: given your access patterns, could use a struct to pack `remainingDailyBorrowLimit ` + `lastDailyBorrowLimitUpdate ` for some gas savings on read/write.

src/BorrowController.sol
```diff
@@ -125,22 +119,39 @@ contract BorrowController {
...
     }
 
+    /**
+     * @notice Returns the available borrow limit of a market
+     * @param market The address of the market to return the available borrow limt
+     */
+    function availableBorrowLimit(address market) public view returns(uint) {
+        uint timeElapsed = block.timestamp - lastDailyBorrowLimitUpdate[market];
+        uint dailyLimit = dailyLimits[market];
+        uint newLimit = remainingDailyBorrowLimit[market] + timeElapsed * dailyLimit / 1 days;
```
wavey0x, line 142: seems like `dailyLimit` should be removed from this formula?

PR #102, opened 2024-12-19
Feat: Pendle PT Helper

0xtj24:
- Add ALE and market Pendle PT Helper
- Add custom ALE for Pendle
- Add fork tests


src/util/PendlePTHelper.sol
```diff
@@ -0,0 +1,281 @@
...
+        uint256 ptBal = pt.balanceOf(address(this)) - ptBalBefore;
+
+        if (ptBal < minMint) revert InsufficientPT();
+        if (recipient != address(this)) pt.safeTransfer(recipient, ptBal);
+        // Send YT to user if specified
+        if (ytRecipient != address(0)) {
+            uint256 ytBalMinted = yt.balanceOf(address(this)) - ytBalBefore;
+            if (ytBalMinted < minMint) revert InsufficientYT();
+            yt.safeTransfer(ytRecipient, ytBalMinted);
+        }
```
08xmt: Should probably fail if `ytRecipient` isn't specified. There shouldn't be a situation where YT tokens are just left in the contract.

08xmt: Would be very useful if the YT could be sold for more collateral, but not sure how realistic that is.

src/util/PendlePTHelper.sol
```diff
@@ -0,0 +1,281 @@
...
+            yt.safeTransferFrom(ytProvider, address(this), amount);
+            yt.approve(router, amount);
+        }
+
+        IERC20 pt = IERC20(markets[market].pt);
+        pt.safeTransferFrom(msg.sender, address(this), amount);
+        pt.approve(router, amount);
+
+        uint256 dolaBal = DOLA.balanceOf(recipient);
+        (bool success, ) = router.call(callData);
```
08xmt: I'm assuming that in cases where the ytProvider is set, the router will know how to just redeem the PT & YT for collateral, while in cases where only PT tokens are provided, it will be able to either:
1. Sell the PT for collateral if before redemption date.
2. Redeem the PT for collateral if after the redemption date.

src/util/PendlePTHelper.sol
```diff
@@ -0,0 +1,281 @@
...
+    ) public override returns (uint256 dolaAmount) {
+        (
+            address market,
+            uint256 minOut, // Minimum amount of DOLA to receive
+            address ytProvider,
+            bytes memory callData
+        ) = abi.decode(data, (address, uint256, address, bytes));
+        _revertIfMarketNotSet(market);
+
+        if (ytProvider != address(0)) {
```
08xmt: Should have a check and error message in case `ytProvider` doesn't have a large enough balance to supply the YT tokens.

08xmt: Thinking about it, how is the `ytProvider` supposed to work? If it's a user address or contract, that's a security vulnerability.
1. Honest user approves YT tokens as `ytProvider` for the helper contract
2. Attacker spots approve in mempool.
3. Attacker deploys PT collateral to market, then withdraws using `transformFromCollateral` with honest user as `ytProvider`

08xmt: Either only allow `ytProvider` and recipient to be same address, or only allow `ytProvider` to be `msg.sender`.

src/util/PendlePTHelper.sol
```diff
@@ -0,0 +1,281 @@
...
+            permit.v,
+            permit.r,
+            permit.s
+        );
+
+        if (ytProvider != address(0)) {
+            IERC20 yt = IERC20(markets[market].yt);
+            yt.safeTransferFrom(ytProvider, address(this), amount);
+            yt.approve(router, amount);
+        }
```
08xmt: Same security problem as previously mentioned

test/util/aleTests/ALEPendlePTUSDe.t.sol
```diff
@@ -0,0 +1,638 @@
...
+    address pendleYT = address(0x96512230bF0Fa4E20Cf02C3e8A7d983132cd2b9F);
+    address marketPT = address(0xcDd26Eb5EB2Ce0f203a84553853667aE69Ca29Ce);
+
+    // Pendle Router Mock
+    address pendleYTHolder =
+        address(0x9844c3688dAaA98De18fBe52499A6B152236896b);
+
+    // Get initial PT
+    // Swap 100K DOLA for PT (recipient helper)
+    bytes pendleDataInitial =
```
08xmt, line 33: MY EYES
image

08xmt: I will just assume this is good and correct

08xmt (changes requested):

src/util/PendlePTHelper.sol
0xtj24: Totally right, we should use `msg.sender`, which is fine for direct interaction but in the case of the ALE we would have to pass another param (`_user`) from the ALE which is the actual `msg.sender`, so for this we would need to deploy an ALE instance for Pendle markets which pass `_user` into the helper and ofc limiting only the ALE to use that function.
At this point we could do the same when minting PT and YT, sending any YT to either `_user` or `msg.sender` depending if it’s via ALE or direct interaction with the helper, so we avoid the possibility of YT tokens left in the helper (related to the previous review comment).

So basically:
- Adapt ALE to pass `_user` into helper call
- Use `_user` or `msg.sender` as YT provider/receiver, limiting functions with `_user` as param to the ALE only.

 Unless you have any better suggestion to avoid a separate ALE for Pendle markets

src/util/PendlePTHelper.sol
0xtj24: The PT tokens can be sold Only before maturity so when requesting the `callData` from Pendle API if it’s after maturity the swap option won’t be available, only redemptions will return a valid `callData`.
Actually after maturity the YT is not needed so I’ll fix it together with the other one to check for maturity and just using PT

src/util/PendlePTHelper.sol
0xtj24: Fixed in [419181f](https://github.com/InverseFinance/FiRM/pull/102/commits/419181fd82b478aaad2453a218e175544b31be1c) and following

src/util/PendlePTHelper.sol
0xtj24: Fixed in [419181f](https://github.com/InverseFinance/FiRM/pull/102/commits/419181fd82b478aaad2453a218e175544b31be1c) and following

src/util/PendlePTHelper.sol
0xtj24: Fixed in [419181f](https://github.com/InverseFinance/FiRM/pull/102/commits/419181fd82b478aaad2453a218e175544b31be1c) and following

src/util/PendlePTHelper.sol
0xtj24: Fixed in [419181f](https://github.com/InverseFinance/FiRM/pull/102/commits/419181fd82b478aaad2453a218e175544b31be1c) and following

src/util/ALEPendle.sol
```diff
@@ -0,0 +1,749 @@
...
+        address collateral,
+        address helper
+    );
+
+    // 1Inch ExchangeProxy address.
+    address payable public exchangeProxy;
+
+    IDBR public constant DBR = IDBR(0xAD038Eb671c44b853887A7E32528FaB35dC5D710);
+
+    IERC3156FlashLender public constant flash =
```
08xmt: Not sure if this should be setable or remain constant.

src/util/ALEPendle.sol
```diff
@@ -0,0 +1,749 @@
...
+    // in other cases it could be different (eg. st-yCRV is collateral, yCRV is the token to be swapped from/to DOLA)
+    // or with DOLA curve LPs, LP token is the collateral and DOLA is the token to be swapped from/to
+    mapping(address => Market) public markets;
+
+    modifier dolaSupplyUnchanged() {
+        uint256 totalSupply = dola.totalSupply();
+        _;
+        if (totalSupply != dola.totalSupply())
+            revert TotalSupplyChanged(totalSupply, dola.totalSupply());
+    }
```
08xmt: This modifier is no longer required with the new flashminter, right?

src/util/ALEPendle.sol
```diff
@@ -0,0 +1,749 @@
...
+            _dbrData.amountIn
+        );
+    }
+
+    /// @notice Mint DOLA to this contract and approve the spender
+    /// @param spender The spender address
+    /// @param _value Amount of DOLA to mint and approve
+    function _approveDola(address spender, uint256 _value) internal {
+        dola.approve(spender, _value);
+    }
```
08xmt: This doesn't have to be an internal function. More readable to write `dola.approve(spender, value)` than calling an internal function to do so.

src/util/ALEPendle.sol
```diff
@@ -0,0 +1,749 @@
...
+        if (totalSupply != dola.totalSupply())
+            revert TotalSupplyChanged(totalSupply, dola.totalSupply());
+    }
+
+    constructor(
+        address _exchangeProxy,
+        address _pool
+    ) Ownable(msg.sender) CurveDBRHelper(_pool) {
+        exchangeProxy = payable(address(_exchangeProxy));
+        _approveDola(address(flash), type(uint).max);
```
08xmt:
```suggestion
        dola.approve(address(flash), type(uint).max);
```

src/util/ALEPendle.sol
```diff
@@ -0,0 +1,749 @@
...
+        if (_helper == address(0)) {
+            if (_buySellToken != IMarket(_market).collateral()) {
+                revert MarketSetupFailed(
+                    _market,
+                    _buySellToken,
+                    IMarket(_market).collateral(),
+                    _helper
+                );
+            }
+        }
```
08xmt: Nit: Avoid unnecessary indentation
```suggestion
        if (_helper == address(0) &&
        _buySellToken != IMarket(_market).collateral()) {
              revert MarketSetupFailed(
                  _market,
                  _buySellToken,
                  IMarket(_market).collateral(),
                  _helper
              );
        }
```

src/util/ALEPendle.sol
```diff
@@ -0,0 +1,749 @@
...
+        address collateral = IMarket(_market).collateral();
+        markets[_market].buySellToken = IERC20(_buySellToken);
+        markets[_market].collateral = IERC20(collateral);
+        markets[_market].buySellToken.approve(_market, type(uint256).max);
+
+        if (_buySellToken != collateral) {
+            markets[_market].collateral.approve(_market, type(uint256).max);
+        }
+
+        if (_helper != address(0)) {
```
08xmt: Will there ever be pendle markets added where there won't be a helper?

src/util/ALEPendle.sol
```diff
@@ -0,0 +1,749 @@
...
+        address spender,
+        bytes calldata swapCallData,
+        Permit calldata permit,
+        bytes calldata helperData,
+        DBRHelper calldata dbrData
+    ) public payable nonReentrant dolaSupplyUnchanged {
+        if (address(markets[market].buySellToken) == address(0))
+            revert MarketNotSet(market);
+
+        bytes memory data = abi.encode(
```
08xmt: Might save some gas on using `abi.encodePacked`. Might even be worth it to have a helper function that does all of the packing so all the encoding can be done off-chain and only the unpacking needs to be done on-chain.

src/util/ALEPendle.sol
```diff
@@ -0,0 +1,749 @@
...
+        IMarket(_market).deposit(
+            _user,
+            markets[_market].collateral.balanceOf(address(this))
+        );
+
+        _borrowDola(_user, _value, _permit, _dbrData, IMarket(_market));
+
+        if (_dbrData.dola != 0) dola.transfer(_user, _dbrData.dola);
+
+        if (_dbrData.amountIn != 0)
```
08xmt: Make sure that if there's any dbr data that points to the user wanting dbr, the _buyDbr function is called.
```suggestion
        if (_dbrData.amountIn > 0 || _dbrData.minOut > 0)
```

src/util/ALEPendle.sol
```diff
@@ -0,0 +1,749 @@
...
+    ) internal {
+        uint256 dolaToBorrow = _value;
+
+        if (_dbrData.dola != 0) {
+            dolaToBorrow += _dbrData.dola;
+        }
+
+        if (_dbrData.amountIn != 0) {
+            dolaToBorrow += _dbrData.amountIn;
+        }
```
08xmt: Nit: Checking for 0 barely saves any gas and makes the code less readable imo.
```suggestion
        uint256 dolaToBorrow = _value + _dbrData.dola + _dbrData.amountIn;
```

src/util/ALEPendle.sol
```diff
@@ -0,0 +1,749 @@
...
+            _permit.v,
+            _permit.r,
+            _permit.s
+        );
+
+        if (dola.balanceOf(address(this)) < dolaToBorrow)
+            revert DOLAInvalidBorrow(
+                dolaToBorrow,
+                dola.balanceOf(address(this))
+            );
```
08xmt: Is this a necessary check?

src/util/ALEPendle.sol
```diff
@@ -0,0 +1,749 @@
...
+        IMarket market
+    ) internal {
+        if (_dbrData.dola != 0) {
+            dola.transferFrom(_user, address(this), _dbrData.dola);
+            _approveDola(address(market), _value + _dbrData.dola);
+            market.repay(_user, _value + _dbrData.dola);
+        } else {
+            _approveDola(address(market), _value);
+            market.repay(_user, _value);
+        }
```
08xmt: Brevity
```suggestion
        if (_dbrData.dola != 0) {
            dola.transferFrom(_user, address(this), _dbrData.dola);
            _value + _dbrData.dola;
        }
        _approveDola(address(market), _value);
        market.repay(_user, _value);
```

src/util/ALEPendle.sol
```diff
@@ -0,0 +1,749 @@
...
+    /// @dev Requires user to sign message to permit the contract to withdraw collateral from the escrow
+    /// @param value Amount of DOLA to repay
+    /// @param market The market contract
+    /// @param collateralAmount Collateral amount to withdraw from the escrow
+    /// @param spender The `allowanceTarget` field from the API response.
+    /// @param swapCallData The `data` field from the API response.
+    /// @param permit Permit data
+    /// @param helperData Optional helper data in case collateral needs to be transformed
+    /// @param dbrData Optional data in case the user wants to sell DBR
+    function deleveragePosition(
```
08xmt: The `leveragePosition` and `deleveragePosition` are almost the same function. If data was fed as a calldata parameter they would be the same function. Probably best to just roll them into one.

src/util/ALEPendle.sol
```diff
@@ -0,0 +1,749 @@
...
+                _helperData
+            );
+            // Reimburse leftover collateral from conversion if any
+            uint256 collateralLeft = markets[_market].collateral.balanceOf(
+                address(this)
+            );
+
+            if (collateralLeft != 0) {
+                markets[_market].collateral.safeTransfer(_user, collateralLeft);
+            }
```
08xmt: In what situations should there be collateral leftover, and how often will that leftover collateral be worth the gas cost to transfer it back to the receiver?

src/util/ALEPendle.sol
```diff
@@ -0,0 +1,749 @@
...
+        // Collateral amount is now converted into sellToken
+        uint256 assetAmount = markets[_market].helper.convertFromCollateral(
+            _user,
+            _collateralAmount,
+            _helperData
+        );
+        uint256 actualAssetAmount = sellToken.balanceOf(address(this));
+
+        if (actualAssetAmount < assetAmount)
+            revert WithdrawFailed(assetAmount, actualAssetAmount);
```
08xmt: Is there a reason we can't trust the amount returned by the helper?

src/util/ALEPendle.sol
```diff
@@ -0,0 +1,749 @@
...
+            _user,
+            _assetAmount,
+            _helperData
+        );
+
+        uint256 actualCollateralAmount = markets[_market].collateral.balanceOf(
+            address(this)
+        );
+        if (actualCollateralAmount < collateralAmount)
+            revert DepositFailed(collateralAmount, actualCollateralAmount);
```
08xmt:
```suggestion
        uint256 actualCollateralAmount = markets[_market].collateral.balanceOf(
            address(this)
        );
        if (actualCollateralAmount < collateralAmount)
            revert DepositFailed(collateralAmount, actualCollateralAmount);
```
Is there a reason we can't trust the amount returned by the helper, and if so, can't we just return the collateral balance?

src/util/PendlePTHelper.sol
```diff
@@ -278,4 +371,77 @@ contract PendlePTHelper is Sweepable, IMultiMarketTransformHelper {
         delete markets[market];
         emit MarketRemoved(market);
     }
+
+    /// UTILS
+    function getSelector(bytes memory _bytes) internal pure returns (bytes4) {
```
08xmt: I don't understand what this function is doing.

src/util/ALEPendle.sol
0xtj24: Fixed in [713b3e4](https://github.com/InverseFinance/FiRM/pull/102/commits/713b3e442a21ede20af90cb75d1bc2ef49a28e82)

src/util/ALEPendle.sol
0xtj24: Fixed in [713b3e4](https://github.com/InverseFinance/FiRM/pull/102/commits/713b3e442a21ede20af90cb75d1bc2ef49a28e82)

src/util/ALEPendle.sol
0xtj24: Fixed in [713b3e4](https://github.com/InverseFinance/FiRM/pull/102/commits/713b3e442a21ede20af90cb75d1bc2ef49a28e82)

src/util/ALEPendle.sol
0xtj24: Fixed in [713b3e4](https://github.com/InverseFinance/FiRM/pull/102/commits/713b3e442a21ede20af90cb75d1bc2ef49a28e82)

src/util/ALEPendle.sol
0xtj24: Fixed in [713b3e4](https://github.com/InverseFinance/FiRM/pull/102/commits/713b3e442a21ede20af90cb75d1bc2ef49a28e82)

src/util/ALEPendle.sol
0xtj24: Fixed in [e8611f5](https://github.com/InverseFinance/FiRM/pull/102/commits/e8611f54da2e64d920168ede36806a5c3622df64)

src/util/ALEPendle.sol
0xtj24: if only `minOut` is set but `amountIn` is zero `buyDBR` won't do anything, but what we could do is to make this check inclusive with `&&` so only if a `minOut` is set there will be a buy, at least we avoid a sandwich

src/util/ALEPendle.sol
0xtj24: Fixed in [1d87b03](https://github.com/InverseFinance/FiRM/pull/102/commits/1d87b03730125558de481d19b3bd46aac1c0e7b6)

src/util/ALEPendle.sol
0xtj24: Fixed in [1d87b03](https://github.com/InverseFinance/FiRM/pull/102/commits/1d87b03730125558de481d19b3bd46aac1c0e7b6)

src/util/ALEPendle.sol
0xtj24: Could probably remain constant, if we were going to upgrade the actual flash minter, the ALE would also possibly require some adjustment too so we will end up deploying a new version, and by being constant we can save some gas

src/util/ALEPendle.sol
0xtj24: Left unchanged after internal discussion and analysis

src/util/ALEPendle.sol
0xtj24: Left unchanged after internal discussion and analysis

src/util/ALEPendle.sol
0xtj24: Left unchanged after internal discussion and analysis

src/util/ALEPendle.sol
0xtj24: Left unchanged after internal discussion and analysis

src/util/ALEPendle.sol
0xtj24: Left unchanged after internal discussion and analysis

src/util/ALEPendle.sol
0xtj24: Left unchanged after internal discussion and analysis

src/util/ALEPendle.sol
```diff
@@ -186,15 +179,16 @@ contract ALEPendle is
...
-            if (_buySellToken != IMarket(_market).collateral()) {
-                revert MarketSetupFailed(
-                    _market,
-                    _buySellToken,
-                    IMarket(_market).collateral(),
-                    _helper
-                );
-            }
+        if (
+            _helper == address(0) &&
```
08xmt: Should we always just revert if `_helper == address(0)` for the pendle ALE?

src/util/ALEPendle.sol
08xmt: > if only `minOut` is set but `amountIn` is zero `buyDBR` won't do anything, but what we could do is to make this check inclusive with `&&` so only if a `minOut` is set there will be a buy, at least we avoid a sandwich

Yeah good point.

08xmt (approved):

src/util/ALEPendle.sol
0xtj24: Fixed in [3cc6e03](https://github.com/InverseFinance/FiRM/pull/102/commits/3cc6e034f7263da8e2eb05aad49fdf2834667239)

wavey0x:
Mismatching logic. 

In `setMarket`, a valid helper is required.
https://github.com/InverseFinance/FiRM/blob/3cc6e034f7263da8e2eb05aad49fdf2834667239/src/util/ALEPendle.sol#L183

But that helper can be cleared to 0x0, as a valid helper is not required during `updateMarketHelper`
https://github.com/InverseFinance/FiRM/blob/3cc6e034f7263da8e2eb05aad49fdf2834667239/src/util/ALEPendle.sol#L229

wavey0x:
optional:
might consider importing the [source library](https://github.com/GNSPS/solidity-bytes-utils/blob/master/contracts/BytesLib.sol#L228) rather than a [modified copy/paste](https://github.com/InverseFinance/FiRM/blob/3cc6e034f7263da8e2eb05aad49fdf2834667239/src/util/PendlePTHelper.sol#L376).

this method should not increase your contract's bytecode size as the library's functions are internal, and the compiler will exclude any bytecode from unused functions.

PR #109, opened 2025-02-13
Feat: Dola-DeUSD markets

0xtj24:
- Add Convex and YearnV2 Market and ALE tests
- Add MockFeedDescription (for the new feed architecture)


0xtj24:
Already merged into `dev` with #102

PR #110, opened 2025-02-19
Feat: sdeUSD market

0xtj24:
- Add sdeUSD feed
- Add feed and market tests


src/feeds/ERC4626Feed.sol
```diff
@@ -0,0 +1,111 @@
...
+            int256 normalizedAssetToUsdPrice,
+            uint256 startedAt,
+            uint256 updatedAt,
+            uint80 answeredInRound
+        ) = feed.latestRoundData();
+
+        uint256 assetToUnderlyingRate;
+
+        try vault.previewRedeem(SCALE) returns (uint256 rate) {
+            assetToUnderlyingRate = rate.divWadDown(assetScale);
```
wavey0x, line 74: this should be correct as it should normalize the rate of a non 18 decimal asset to an 18 decimal rate.

so in example a token with 6 decimals: 1e6 * 1e18 / 1e6 = 1e18

wavey0x (commented):
LGTM

PR #111, opened 2025-03-03
Fix: Pectra + DBR bug low debt

08xmt:
Documentation for `msgSender.code.length == 0`
image


src/BorrowController.sol
```diff
@@ -120,7 +121,7 @@ contract BorrowController {
         //If the chainlink oracle price feed is stale, deny borrow
         if(isPriceStale(msg.sender)) return false;
         //If the message sender is not a contract, then there's no need check allowlist
-        if(msgSender == tx.origin) return true;
+        if(msgSender == tx.origin && msgSender.code.length == 0) return true;
```
nourharidy, line 118: Can't this be bypassed using a constructor?

PR #113, opened 2025-03-12
Feat: USDe before maturity feed with NAV

0xtj24:
- Add tests to use NAV instead of Fixed Price
- Add USDe before maturity feed with NAV and tests

### DIFF
--- a/src/feeds/USDeNavBeforeMaturityFeed.sol
+++ b/src/feeds/USDeNavBeforeMaturityFeed.sol
@@ -6,8 +6,7 @@ import {IERC4626} from "lib/openzeppelin-contracts/contracts/interfaces/IERC4626
 
 interface INavFeed {
     function maturity() external view returns (uint256);
-    function getDiscount(uint256 timeLeft) external view returns (uint256);
-
+    function decimals() external view returns (uint8);
     function latestRoundData()
         external
         view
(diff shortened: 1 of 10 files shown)


src/feeds/USDeNavBeforeMaturityFeed.sol
```diff
@@ -0,0 +1,88 @@
...
+    {
+        (
+            uint80 roundId,
+            int256 sUSDePrice,
+            uint startedAt,
+            uint updatedAt,
+            uint80 answeredInRound
+        ) = sUSDeFeed.latestRoundData();
+
+        uint256 sUSDeToUSDeRate = sUSDe.convertToAssets(1e18);
```
wavey0x, line 61: if token is capable of taking withdraw fees, `.previewRedeem()` may be more accurate

wavey0x: also, i think this particular token is subject to a cooldown. 

thus, for the goal of accurate pricing for sake of lending market risk, the value of staked vs unstaked version actually deviate from simple share price formula.

just wanted to highlight this note, though im sure you've already considered this.

src/feeds/USDeNavBeforeMaturityFeed.sol
```diff
@@ -0,0 +1,88 @@
...
+
+        // divide sUSDe/USD by sUSDe/USDe rate to get USDe/USD price
+        int256 USDeUsdPrice = (sUSDePrice * 1e18) / int256(sUSDeToUSDeRate);
+
+        (,int256 navDiscountedPrice,,,)= navFeed.latestRoundData();
+        int256 usdeDiscountPrice = (USDeUsdPrice * navDiscountedPrice) / 1e18;
```
wavey0x, line 67: i believe this particular token is subject to a cooldown.

thus, a simple share price formula may be a somewhat misleading way to get the actual market price as the market should theoretically apply a discount based on time it takes to cool down.

just wanted to highlight this note, though i'm sure you've already considered.

wavey0x (approved):

wavey0x (approved):
LGTM

0xtj24:
Already merged into `dev` with #102

PR #130, opened 2025-11-06
Feat/dynamic fee curve feed

08xmt:


src/interfaces/ICurvePool.sol
```diff
@@ -77,4 +77,8 @@ interface ICurvePool {
     function lp_token() external view returns (address);
 
     function decimals() external view returns (uint256);
+
+    function fee() external view returns(uint);
+
+    function out_fee() external view returns(uint);
```
0xtj24, line 83: Unused

src/feeds/DynamicFeeCurveFeed.sol
```diff
@@ -0,0 +1,126 @@
...
+     * @return int256 Returns the last finalized price of the chainlink oracle
+     */
+    function latestAnswer() external view returns (int256) {
+        (, int256 latestPrice, , , ) = latestRoundData();
+        return latestPrice;
+    }
+
+    function setMaxFee(int _maxFee) external {
+        require(msg.sender == gov, "ONLY GOV");
+        require(_maxFee <= 1e10, "CurveFeed: maxFee > 100%");
```
wavey0x: suggestion: edit require statement to prevent negative values
`require(_maxFee >= 0 && _maxFee <= 1e10)`

if ever a negative value is provided, it will cause this line to always clamp to it, which i presume is not intentional.
https://github.com/InverseFinance/FiRM/blob/c15d14b09251951f5762a57feb4d4c5b14aacfb0/src/feeds/DynamicFeeCurveFeed.sol#L86

wavey0x (approved):

src/feeds/DynamicFeeCurveFeed.sol
08xmt: Fixed in https://github.com/InverseFinance/FiRM/pull/130/commits/f1628fceb4ea81ec35e026e9570ce8ebf3cf7402

PR #134, opened 2026-01-10
Feat: FeedSwitch V2

0xtj24:
- Add FeedSwitchV2

### DIFF
--- a/src/util/FeedSwitchV2.sol
+++ b/src/util/FeedSwitchV2.sol
@@ -1,0 +1,163 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.18;
+
+import "src/interfaces/IChainlinkFeed.sol";
+
+
+/// @title FeedSwitch
+/// @notice A contract to switch between feeds after a timelock period
+/// @dev The switch can only be initiated by the guardian and will be effective after the timelock period has passed, which could be zero.
+/// If the switch is initiated but not yet effective, it can be cancelled by the guardian without timelock period (if is not zero).
+/// The guardian can initiate the switch again to the previous feed multiple times.
+contract FeedSwitchV2 {
+    error NotGuardian();
+    error NotGov();
+    error NotPendingGov();
+    error FeedDecimalsMismatch();
+
+    address public pendingGov;
+    address public gov;
+    uint256 public timelockPeriod;
+    IChainlinkFeed public feed;
+    IChainlinkFeed public previousFeed;
+
+    uint256 public switchCompletedAt;
+
+    mapping(address => bool) public isGuardian;
+
+    IChainlinkFeed public immutable initialFeed;
+    IChainlinkFeed public immutable fallbackFeed;
+
+    event FeedSwitchInitiated(address indexed newFeed);
+    event NewPendingGov(address indexed pendingGov);
+    event GovChanged(address indexed newGov);
+    event GuardianSet(address indexed guardian, bool isGuardian);
+    event TimelockPeriodChanged(uint256 newTimelockPeriod);
+
+    constructor(
+        address _initialFeed,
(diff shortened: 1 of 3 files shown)


src/util/FeedSwitchV2.sol
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    modifier onlyGov() {
+        if (msg.sender != gov) revert NotGov();
+        _;
+    }
+
+    /// @notice Initiate the feed switch, entering the timelock period
+    /// @dev Can only be called by the guardian
+    function initiateFeedSwitch() external {
```
08xmt: Make clear function can switch between feeds.
```suggestion
    function toggleFeedSwitch() external {
```

src/util/FeedSwitchV2.sol
```diff
@@ -0,0 +1,163 @@
...
+        return 18;
+    }
+
+    /// @notice Check if the feed switch is queued
+    /// @dev If not queued, will return false and 0 as time left
+    /// @return isQueued Whether the feed switch is queued
+    /// @return timeLeft The time left for the switch to be effective
+    function isFeedSwitchQueued() external view returns (bool, uint256) {
+        bool isQueued = switchCompletedAt > 0 &&
+            block.timestamp < switchCompletedAt;
```
08xmt: Can be simplified:
```suggestion
        bool isQueued = block.timestamp < switchCompletedAt;
```

src/util/FeedSwitchV2.sol
```diff
@@ -0,0 +1,163 @@
...
+    /// @return decimals The number of decimals
+    function decimals() external pure returns (uint8) {
+        return 18;
+    }
+
+    /// @notice Check if the feed switch is queued
+    /// @dev If not queued, will return false and 0 as time left
+    /// @return isQueued Whether the feed switch is queued
+    /// @return timeLeft The time left for the switch to be effective
+    function isFeedSwitchQueued() external view returns (bool, uint256) {
```
08xmt: Can simplify return type to either be 0 or time of switch, but not important.

src/util/FeedSwitchV2.sol
0xtj24: Fixed in [7699f12](https://github.com/InverseFinance/FiRM/pull/134/commits/7699f1272feb50c21bbd6c16a12e6fe6097883cc)

src/util/FeedSwitchV2.sol
0xtj24: Fixed in [7699f12](https://github.com/InverseFinance/FiRM/pull/134/commits/7699f1272feb50c21bbd6c16a12e6fe6097883cc)

src/util/FeedSwitchV2.sol
0xtj24: Fixed in [7699f12](https://github.com/InverseFinance/FiRM/pull/134/commits/7699f1272feb50c21bbd6c16a12e6fe6097883cc)

src/util/FeedSwitchV2.sol
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    modifier onlyGov() {
+        if (msg.sender != gov) revert NotGov();
+        _;
+    }
+
+    /// @notice Toggle the feed switch, entering or exiting the timelock period
+    /// @dev Can only be called by the guardian
+    function toggleFeedSwitch() external {
```
wavey0x, line 64: note:
current implentation is functionally correct, but semantics differ from conventional timelocks which may lead to confusion for integrators 
- toggle function is used both to queue a switch and cancel an in-flight switch
- the feed storage variable is updated immediately, may lead to confusion for integrators as which is the active
- `FeedSwitchInitiated` is emitted even on cancelations

PR #141, opened 2026-07-08
Add ad-hoc token approvals for all external calls in ALE

08xmt:


08xmt:
https://v12.sh/runs/4735/public

08xmt:
## Findings

### NM-001 — LOW — `msg.value` is always 0 inside the flash callback; ETH-fee swap path is dead plumbing  — ✅ FIXED
> **Resolution:** `msg.value` is now captured in the payable entrypoints (`leveragePosition`
> `:253`, `deleveragePosition` `:352`), threaded through the ABI `data` payload as a trailing
> `uint256`, decoded in the callback (`_ethValue`), and forwarded from the contract's own ETH
> balance via `call{value: _ethValue}(...)` in both `_onFlashLoan*` functions. The naive
> `flash.flashLoan{value: msg.value}` was deliberately avoided (the ERC-3156 lender does not
> forward ETH into the callback). `depositAndLeveragePosition` works too: it calls
> `leveragePosition` internally, so `msg.value` is inherited. `forge build` passes.

**Source:** Feynman Cat. 2/7 (ordering / external-call context)
**Where:** `leveragePosition` (254), `_onFlashLoanLeverage` (420), `_onFlashLoanDeleverage` (535), `_refundExcess` (712–713)

`leveragePosition` / `deleveragePosition` are `payable` and call
`payable(_proxy).call{value: msg.value}(_swapCallData)` inside the callback to fund an
on-chain protocol fee. But:
- `flash.flashLoan(...)` is invoked **without** forwarding value (no `{value:}`), and the DOLA
  flash-minter does not forward ETH, so **`msg.value == 0` throughout `onFlashLoan`**.
- Any ETH the user sends is not used for the swap; it sits in the contract and is
  round-tripped back to the user by `_refundExcess` (`address(this).balance` → `_user`).

**Impact:** The ETH-fee feature is non-functional. A 0x quote that requires an attached ETH
protocol fee would revert. No fund loss (ETH is refunded). For ERC20↔ERC20 0x v4 swaps
(the actual use case) this is inert.
**Fix:** Either forward `msg.value` into `flash.flashLoan{value: msg.value}(...)` **and**
confirm the lender forwards it to the callback, or drop the `payable` / `{value:}` plumbing to
avoid the false affordance.

### NM-002 — LOW — DBR-buy with `minOut == 0` silently over-borrows instead of buying DBR
**Source:** Feynman Cat. 6 (silent fallthrough) → State (borrow vs spend mismatch)
**Where:** `_borrowDola` (587) vs `_onFlashLoanLeverage` (453–454)

`_borrowDola` always borrows `dolaToBorrow = _value + _dbrData.dola + _dbrData.amountIn`.
The DBR purchase, however, is guarded by **`amountIn > 0 && minOut > 0`** (line 453). With
`amountIn > 0, minOut == 0`, the contract borrows the extra `amountIn` DOLA but **skips
`_buyDbr`**; the unused DOLA is then handed back to the user by `_refundExcess`.

**Impact:** User silently takes on `amountIn` of extra debt and receives `amountIn` DOLA in
wallet instead of the DBR they intended (no DBR coverage bought). No protocol loss; a UX
footgun that only bites a caller who already passed an unprotected `minOut == 0`. Symmetric
`_sellDbr` guard in deleverage (558) is a no-op in the same case (nothing borrowed), so it is
harmless there.
**Fix:** Gate the extra borrow on the same condition, i.e. only add `amountIn` to
`dolaToBorrow` when `minOut > 0` (or revert on `amountIn > 0 && minOut == 0`).

### NM-003 — LOW — Leftover `buySellToken` after helper conversion is not swept on the leverage path
**Source:** State parallel-path comparison (leverage vs deleverage asymmetry)
**Where:** `_onFlashLoanLeverage` (433–447) vs `_onFlashLoanDeleverage` (517–556)

Deleverage sweeps residual `collateral` (518–524) **and** residual `sellToken` (552–556)
back to the user. Leverage deposits the full `collateral` balance but **never sweeps residual
`buySellToken`** left behind if `helper.convertToCollateral` consumes less than approved. Any
such dust is stranded in the contract and, because every entry point reads
`buySellToken.balanceOf(address(this))` (427), is folded into the **next** caller's position.

**Impact:** Small value leak for the depositing user; inconsistent with the deleverage path.
Not a theft primitive (no standing approvals; requires a helper that under-consumes).
**Fix:** After the deposit, if `buySellToken != collateral`, sweep any residual
`buySellToken` to `_user`, mirroring deleverage.

### NM-004 — LOW — Raw `.approve` on arbitrary collateral / buySell tokens breaks for non-bool-returning ERC20s (e.g. USDT)  — ✅ FIXED
> **Resolution:** switched the four collateral/buySell approvals (446, 533-534, 652, 683) to
> `SafeERC20.forceApprove`; the deleverage `approve(_proxy,0)`+`approve(_proxy,amount)` pair
> collapsed to a single `forceApprove` (reset-to-zero is handled internally). DOLA approvals
> keep the local-`IERC20` raw `.approve` (bool-returning, safe). Applied to both file copies;
> `forge build` passes.

**Source:** Feynman Cat. 3 (consistency: `SafeERC20` imported but not used for approvals)
**Where:** lines 446, 533–534, 652, 683 (also `CurveHelper` 42/54 for DOLA/DBR — safe)

`using SafeERC20 for IERC20` is declared and used for transfers, but **approvals use the raw
OZ `IERC20.approve` (declared `returns (bool)`)**. Tokens that return no data on `approve`
(USDT and friends) cause the ABI decoder to revert. Note the deleverage swap even applies the
USDT-style **reset-to-zero** pattern (533–534) — anticipating USDT-like allowance semantics —
yet the same call would still revert on a genuine no-return-value USDT because of the missing
`bool`.

**Impact:** Any market whose `collateral` or `buySellToken` is a non-standard ERC20 cannot be
leveraged/deleveraged — every flow reverts. Latent compatibility landmine; caught at market
setup/testing, so no live-fund risk with the current bool-returning token set.
**Fix:** Use `forceApprove` (OZ SafeERC20) for the collateral/buySell approvals; it also makes
the manual `approve(_proxy, 0)` reset redundant.

### NM-005 — INFO — Hard coupling to a **zero-fee** flash minter
**Where:** `leveragePosition` (252), `deleveragePosition` (348), `onFlashLoan` (362 fee ignored), `_refundExcess` (708)

The approval to `flash` is for `value` (not `value + fee`) and `_refundExcess` only guarantees
`_value`. If the hardcoded DOLA flash-minter (`0x6C5F…`) ever charged a non-zero fee, **every**
call would revert on the lender's `transferFrom(value + fee)`. Safe-fail, but a silent
dependency worth a comment/assert.

### NM-006 — INFO — `_refundExcess` refunds ETH with `.transfer` (2300 gas)
**Where:** `_refundExcess` (712–713)

A contract-user that cannot accept ETH within 2300 gas — or any user, if an attacker
`selfdestruct`-feeds dust ETH so `address(this).balance > 0` on a zero-ETH call — will have its
leverage/deleverage revert. Griefing/compat only.
**Fix:** Use a low-level `call` with success check (consistent with the WETH-unwrap path in
`OffchainAbstractHelper`).

### NM-007 — INFO (verified safe) — `updateMarketHelper` no longer revokes the old helper's approval
The pre-refactor code reset old-helper allowances to 0. Under the new ad-hoc model there are no
standing approvals to revoke (each `_convertTo*` approves the exact amount immediately
consumed), so the removal is correct. Noted to confirm it was intentional, not an omission.

### NM-008 — LOW (observability) — `LeverageUp.dolaBorrowed` under-reported the on-behalf borrow  — ✅ FIXED
**Source:** external reviewer (post-audit)
**Where:** `LeverageUp` event (`:116`), emit in `_onFlashLoanLeverage`, `_borrowDola` (`:599`)

The 5th event field is documented as *"amount of DOLA borrowed on behalf of the user"* but the
emit passed `_dbrData.dola` — only the optional extra handed to the user. The amount actually
sent to `borrowOnBehalf` is `_value + _dbrData.dola + _dbrData.amountIn`, so indexers/monitoring
undercounted the debt taken on and couldn't see the DBR-purchase borrow. Value flow unaffected.
`LeverageDown`'s symmetric field (`dolaUserRepaid = _dbrData.dola`) is **correct** and was left
as-is (there `dola` genuinely is the user's own repay contribution).
> **Resolution:** `_borrowDola` now returns its `dolaToBorrow` total (single source of truth);
> `_onFlashLoanLeverage` captures it as `dolaBorrowed` and emits that. The deposit was moved into
> a block scope so its local frees before the emit, resolving a stack-too-deep the fix otherwise
> triggered (project compiles without via-IR). `forge build` passes. **Note:** this changes a
> value off-chain consumers ingest — coordinate with whoever runs the indexer on rollout.

### NM-009 — LOW (defense-in-depth) — `depositAndLeveragePosition` is not `nonReentrant` and holds funds before the lock
**Source:** Pass-2 Feynman Cat. 3/5 (guard consistency + funds-held-before-lock)
**Where:** `depositAndLeveragePosition` (`:288`, `external payable` — no `nonReentrant`)

Unlike `leveragePosition` (`:235`) and `deleveragePosition` (`:336`), this entrypoint has **no
`nonReentrant` guard**. It pulls `initialDeposit` via `safeTransferFrom` (`:299`) — and now also
holds the caller's attached ETH — **before** the reentrancy lock is acquired (the lock is only
taken when it internally calls `leveragePosition`). If a market's `collateral`/`buySellToken` is
a hook-bearing token (ERC-777 / callback-on-transfer), the transfer can re-enter the ALE while
the outer deposit (and ETH) sit in the contract.

**Exploitability: could not construct a profitable path — reporting as hardening.** A reentrant
`leveragePosition` would sweep the outer deposit (`balanceOf(this)`) into the re-entrant
position, but every `borrowOnBehalf` is bounded by the market's credit limit, every flash loan
must be repaid (`_refundExcess`), the ALE holds no *other* user's funds between txs, and the ETH
path self-reverts (the outer `call{value: _ethValue}` fails once the reentrant refund has
drained the balance). Net: the attacker only shuffles their own funds between their own escrows.
Also gated by gov choosing a hook-capable token. Pre-existing (not introduced by the fixes), but
the NM-001 change puts ETH in the contract before the lock, raising its salience.
**Fix:** add `nonReentrant` to `depositAndLeveragePosition` (cheap; matches the sibling
entrypoints and follows checks-effects on the fund pull).

nourharidy (approved):

PR #143, opened 2026-09-21
Add wind-down feed for Curve coin-0 pricing

cryptoharry:
Curve StableSwap's capped EMA creates a downside price floor when inverted to price `coins[0]`. This adds standalone `ChainlinkCurveWindDownFeed` under `src/feeds`, using the existing coin-0 pricing formula with permissionless wind-down and RWG-controlled recovery. `ChainlinkCurveFeed` is unchanged.

### Behaviour

- Constructor parameters: 18-decimal base USD feed, Curve pool, reference oracle index, wind-down duration in seconds and RWG multisig address. The priced asset is always `coins[0]`.
- `canStartWindDown()` checks only that no wind-down is active and EMA >= `1.9e18`. There is no persistence window, additional EMA ceiling or USD timestamp restriction.
- Permissionless `startWindDown()` records the current positive USD price and activation time. The recorded price decays linearly to one raw 18-decimal feed unit over the configured duration, then remains there until RWG resets it.
- While wind-down is active, timestamps and round IDs are zero and neither price source is consulted. Zero timestamps block borrowing where FiRM's borrowing staleness check is enabled.
- Each activation immediately transfers the feed's entire Ethereum mainnet DOLA balance to its caller, including a zero-value transfer when unfunded. `WindDownStarted` records activation and the reward.

Eligibility does not guarantee transaction success: activation still needs a readable positive USD starting price and a successful DOLA call. Stale USD timestamps are accepted.

### RWG recovery

`stopWindDown()` is restricted to the nonzero RWG address fixed at deployment. It clears the recorded starting price and time, restores live pricing and upstream timestamps, and emits `WindDownStopped`. It can also be called after the terminal price is reached.

- Restoring live timestamps can allow borrowing again, subject to normal market checks. Stale upstream data remains stale.
- If EMA remains >= `1.9e18`, anyone can immediately start another wind-down. There is no additional pause mechanism.
- Previous DOLA rewards are not refunded. A subsequent activation pays the then-current balance. There is no separate rescue function.
- Stopping does not undo completed liquidations or clear FiRM Oracle's recorded daily lows. Those lows can continue depressing collateral valuation after live pricing returns.

### Validation

27 local Foundry tests passed, including 512 fuzz cases. Coverage includes normal pricing, activation and decay, rewards, RWG authorization, restoration of live data, repeat activation, recorded-low persistence and expiry, and LP/Yearn timestamp propagation through Oracle and BorrowController.

```sh
forge test --match-path test/feeds/ChainlinkCurveWindDownFeed.t.sol --evm-version shanghai -vv
```

Draft for review; no deployment has been performed.


src/feeds/ChainlinkCurveWindDownFeed.sol
```diff
@@ -0,0 +1,137 @@
...
+///      The RWG address fixed at deployment can stop wind-down and restore live pricing and timestamps.
+///      Fund this feed with Ethereum mainnet DOLA before activation to offer a caller reward.
+///      The caller receives the full balance in the activation transaction, including a zero transfer.
+///      There is no rescue function; DOLA is paid only when a wind-down is started.
+///      Consuming FiRM markets must enable the borrow controller's staleness check.
+contract ChainlinkCurveWindDownFeed {
+    uint256 public constant WIND_DOWN_TRIGGER_EMA = 1.9e18;
+    uint256 public constant TERMINAL_PRICE = 1;
+    uint256 public constant targetIndex = 0;
+    IERC20 public constant dola = IERC20(0x865377367054516e17014CcdED1e7d814EDC9ce4);
```
08xmt: Should be immutable constructor arguments to make feed reusable for other tokens. Probably only necessary to have one of them be set in the constructor, and then the other read during construction from the pool.

src/feeds/ChainlinkCurveWindDownFeed.sol
```diff
@@ -0,0 +1,137 @@
...
+///      Use ChainlinkCurveFeed for nonzero target indices; their direct EMA has no downside floor.
+///      Before activation: base USD price * 1e18 / Curve EMA.
+///      After activation: fixed starting USD price decays to 1 raw feed unit; timestamps are zero.
+///      Activation is permissionless at EMA >= 1.9e18, without a persistence window or timestamp check.
+///      The RWG address fixed at deployment can stop wind-down and restore live pricing and timestamps.
+///      Fund this feed with Ethereum mainnet DOLA before activation to offer a caller reward.
+///      The caller receives the full balance in the activation transaction, including a zero transfer.
+///      There is no rescue function; DOLA is paid only when a wind-down is started.
+///      Consuming FiRM markets must enable the borrow controller's staleness check.
+contract ChainlinkCurveWindDownFeed {
```
08xmt: Nit: Use ALL_CAPS_SNAKECASE for all immutable and constant variables

src/feeds/ChainlinkCurveWindDownFeed.sol
```diff
@@ -0,0 +1,137 @@
...
+            if (elapsed < windDownDuration) {
+                // Safe: activation EMA >= 1.9e18 and checked signed pricing bounds the start price;
+                // the remaining duration is at most uint32.max.
+                price += (windDownStartPrice - TERMINAL_PRICE) * (windDownDuration - elapsed) / windDownDuration;
+            }
+            return (0, int256(price), 0, 0, 0);
+        }
+        int256 assetToUsdPrice;
+        (roundId, assetToUsdPrice, startedAt, updatedAt, answeredInRound) = assetToUsd.latestRoundData();
+        // Same coin-0 calculation as ChainlinkCurveFeed; Solidity checks signed overflow.
```
08xmt: Nit: Unnecessary comment
```suggestion
```

src/feeds/ChainlinkCurveWindDownFeed.sol
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    function latestRoundData()
+        public
+        view
+        returns (uint80 roundId, int256 usdPrice, uint256 startedAt, uint256 updatedAt, uint80 answeredInRound)
+    {
+        if (windDownStartPrice != 0) {
+            // Zero timestamps intentionally signal unusable-for-borrowing data to FiRM.
+            // No upstream calls: outages and apparent recoveries cannot interrupt decay.
```
08xmt: Nit: Unnecessary comment
```suggestion
```

src/feeds/ChainlinkCurveWindDownFeed.sol
```diff
@@ -0,0 +1,137 @@
...
+            uint256 elapsed = block.timestamp - windDownStartedAt;
+            uint256 price = TERMINAL_PRICE;
+            if (elapsed < windDownDuration) {
+                // Safe: activation EMA >= 1.9e18 and checked signed pricing bounds the start price;
+                // the remaining duration is at most uint32.max.
```
08xmt: Nit: Unnecessary comments relating to reverts on overflow. Better to describe what's happening.
```suggestion
//Linear decay towards TERMINAL_PRICE from windDownStartPrice over windDownDuration
```

src/feeds/ChainlinkCurveWindDownFeed.sol
```diff
@@ -0,0 +1,137 @@
...
+                price += (windDownStartPrice - TERMINAL_PRICE) * (windDownDuration - elapsed) / windDownDuration;
+            }
+            return (0, int256(price), 0, 0, 0);
+        }
+        int256 assetToUsdPrice;
+        (roundId, assetToUsdPrice, startedAt, updatedAt, answeredInRound) = assetToUsd.latestRoundData();
+        // Same coin-0 calculation as ChainlinkCurveFeed; Solidity checks signed overflow.
+        usdPrice = (assetToUsdPrice * int256(10 ** decimals())) / int256(curvePool.price_oracle(assetOrTargetK));
+        if (usdPrice <= 0) revert InvalidBasePrice();
```
08xmt: Price feeds should never block on `latestRoundData` calls. Better to return TERMINAL_PRICE with 0'ed updated at, than to block execution if `usdPrice` is zero or below.
```suggestion
        if (usdPrice <= 0) return (0, TERMINAL_PRICE, 0, 0, 0);
```

08xmt (changes requested):

src/feeds/ChainlinkCurveWindDownFeed.sol
cryptoharry: I've kept this specific to coin 0 because it addresses the downside floor created by inverting the capped EMA. Coin 1 can still track a deep depeg, so the existing feed remains appropriate there. Different coin-0 assets are already supported through the pool and base-feed constructor arguments.

DOLA is only the activation reward token, not the asset being priced, so its mainnet address remains fixed.

src/feeds/ChainlinkCurveWindDownFeed.sol
cryptoharry: Done — all constants and immutables now use ALL_CAPS_SNAKECASE.

src/feeds/ChainlinkCurveWindDownFeed.sol
cryptoharry: Removed.

src/feeds/ChainlinkCurveWindDownFeed.sol
cryptoharry: Removed.

src/feeds/ChainlinkCurveWindDownFeed.sol
cryptoharry: Updated to describe the linear decay.

src/feeds/ChainlinkCurveWindDownFeed.sol
cryptoharry: I've changed this to return (0, 0, 0, 0, 0) for a nonpositive calculated price.

Returning 1 could turn an unexpected invalid reading into a near-zero collateral valuation and allow liquidations without the intended gradual wind-down. Returning zero lets FiRM's Oracle reject the invalid price instead.

